PR #84, opened 2020-01-21
Feature/refactor jwt

Nesty92:
Allow to use customs authorization mechanisms in node creation, by default use jwt.


Nesty92:
Fix #80

FZambia:
@Nesty92 thanks a lot! I'll come with review comments very soon - the idea is right, but I have some thoughts about way to structure and name things.

BTW do you need this for some specific use case or simply decided to contribute after looking at issues?

Nesty92:
@FZambia I used some centrifuge ideas to do another project, but in this PR i simply decided to contribute after analyzing the problem, i await the comments of the review

I hope to continue contributing to other issues in the future. :)

errors.go
```diff
@@ -66,4 +66,14 @@ var (
 		Code:    110,
 		Message: "expired",
 	}
+	// ErrTokenInvalid ...
```
FZambia: Not what I suggested actually - sorry that I explained bad:) This is a list of errors that can be returned to client in client-server protocol. If you can see they all start with `Error...`, not from `Err...`. The idea was to introduce new internal errors in `token_verifier.go` file:
```
ErrTokenInvalid = errors.New("...")
ErrTokenInvalidInfo = errors.New("...")
ErrTokenExpired = errors.New("...")
``` 

And use them.

BTW yesterday I realized that as you personally don't need this feature yourself then we can simply do refactoring but not export this API publicly. This will give us more freedom to change things in future without breaking users code. I suppose we already solved an original case why this JWT issue was raised (support RSA tokens) so we can safely wait for other request from library user that will need this as part of public API.  

So what I suggest is make all these changes private for now. This is a wonderful refactoring that I really like but I think if we keep a possibility to not extend public API at all then we must use this chance until there will be request from user. This means that we can keep these new errors private for now:

```
errTokenInvalid = errors.New("...")
errTokenInvalidInfo = errors.New("...")
errTokenExpired = errors.New("...")
```

The same for all other entities in pull request:

```
TokenVerifier
ConnectToken
SubscribeToken
```

Then it will be very simple to make this API public.

token_verifier_jwt.go
```diff
@@ -0,0 +1,119 @@
...
+	"github.com/dgrijalva/jwt-go"
+)
+
+type tokenVerifierJWT struct {
+	mu                 sync.RWMutex
+	TokenHMACSecretKey string
+	TokenRSAPublicKey  *rsa.PublicKey
+}
+
+func NewTokenVerifierJWT(tokenHMACSecretKey string, tokenRSAPublicKey *rsa.PublicKey) tokenVerifier {
```
FZambia: please make this private too

config.go
```diff
@@ -85,6 +85,10 @@ type Config struct {
 	LogLevel LogLevel
 	// LogHandler is a handler func node will send logs to.
 	LogHandler LogHandler
+
+	// tokenVerifier is a interface to allow customs tokens verifiers
+	// Zero value means that TokenVerifierJWT will be used.
+	TokenVerifier tokenVerifier
```
FZambia: I think there is no need to put this into config at moment if we do not allow to override it

token_verifier_jwt_test.go
```diff
@@ -0,0 +1,166 @@
...
+	}{
+		{
+			name:     "",
+			verifier: nil,
+			args:     args{},
+		},
+	}
+	for _, tt := range tests {
+		t.Run(tt.name, func(t *testing.T) {
+			//verifier := &tokenVerifierJWT{
```
FZambia: Do we need this?

client.go
```diff
@@ -1326,79 +1292,43 @@ func (c *Client) refreshCmd(cmd *protocol.RefreshRequest) (*clientproto.RefreshR
...
-			c.node.logger.log(newLogEntry(LogLevelInfo, "invalid refresh token", map[string]interface{}{"error": err.Error(), "client": c.uid, "user": c.UserID()}))
-			return resp, DisconnectInvalidToken
-		}
-		c.node.logger.log(newLogEntry(LogLevelInfo, "invalid refresh token", map[string]interface{}{"error": err.Error(), "client": c.uid, "user": c.UserID()}))
-		return resp, DisconnectInvalidToken
+	var (
+		token     connectToken
+		errVerify error
+	)
+	if token, errVerify = c.node.verifyConnectToken(cmd.Token); errVerify != nil {
```
FZambia: This behaviour does not mimic previous one, no proper token verify error handling. Seems that this is also true for other places

FZambia:
@Nesty92 thanks a lot, looks very good! I will look more carefully on `client.go` during weekend and if won't notice sth then merge it.

Nesty92:
@FZambia great, let me know if you see something

client.go
```diff
@@ -1187,61 +1186,28 @@ func (c *Client) connectCmd(cmd *protocol.ConnectRequest) (*clientproto.ConnectR
...
-					resp.Error = ErrorTokenExpired
-					return resp, nil
-				}
-				c.node.logger.log(newLogEntry(LogLevelInfo, "invalid connection token", map[string]interface{}{"error": err.Error(), "client": c.uid}))
-				return resp, DisconnectInvalidToken
-			}
+		var token connectToken
+		var err error
+		if token, err = c.node.verifyConnectToken(cmd.Token); err != nil {
 			c.node.logger.log(newLogEntry(LogLevelInfo, "invalid connection token", map[string]interface{}{"error": err.Error(), "client": c.uid}))
```
FZambia: move log line below errTokenExpired check - no need to log expired case here

client.go
```diff
@@ -1326,79 +1292,48 @@ func (c *Client) refreshCmd(cmd *protocol.RefreshRequest) (*clientproto.RefreshR
...
 	}
 
 	config := c.node.Config()
 
-	var user string
-	var expireAt int64
-	var info Raw
-	var b64info string
-
-	if cmd.Token == "" {
```
FZambia: looks like previously we had duplicate checks here with different return disconnect codes:) Thanks for fixing.

client.go
```diff
@@ -1326,79 +1292,48 @@ func (c *Client) refreshCmd(cmd *protocol.RefreshRequest) (*clientproto.RefreshR
...
-				return resp, nil
-			}
-			c.node.logger.log(newLogEntry(LogLevelInfo, "invalid refresh token", map[string]interface{}{"error": err.Error(), "client": c.uid, "user": c.UserID()}))
-			return resp, DisconnectInvalidToken
+	var (
+		token     connectToken
+		errVerify error
+	)
+	if token, errVerify = c.node.verifyConnectToken(cmd.Token); errVerify != nil {
+		c.node.logger.log(newLogEntry(LogLevelInfo, "invalid refresh token", map[string]interface{}{"error": errVerify.Error(), "client": c.uid}))
```
FZambia: No need to log when the reason is token expiration

client.go
```diff
@@ -1484,66 +1419,39 @@ func (c *Client) subscribeCmd(cmd *protocol.SubscribeRequest, rw *replyWriter) *
...
-				c.node.logger.log(newLogEntry(LogLevelInfo, "invalid subscription token", map[string]interface{}{"error": err.Error(), "client": c.uid, "user": c.UserID()}))
-				rw.write(&protocol.Reply{Error: ErrorPermissionDenied})
-				return nil
-			}
-			c.node.logger.log(newLogEntry(LogLevelInfo, "invalid subscription token", map[string]interface{}{"error": err.Error(), "client": c.uid, "user": c.UserID()}))
+			c.node.logger.log(newLogEntry(LogLevelInfo, "invalid subscription token", map[string]interface{}{"error": errVerify.Error(), "client": c.uid, "user": c.UserID()}))
 			rw.write(&protocol.Reply{Error: ErrorPermissionDenied})
 			return nil
 		}
+		if c.uid != token.UserID {
```
FZambia: Major thing here: subscribeToken does not use `UserID` - it contains `Client` string - which is an unique id of connection generated on server side after connection established.

client.go
```diff
@@ -1835,94 +1743,58 @@ func (c *Client) subRefreshCmd(cmd *protocol.SubRefreshRequest) (*clientproto.Su
...
 			return resp, nil
 		}
-		c.node.logger.log(newLogEntry(LogLevelInfo, "invalid subscription refresh token", map[string]interface{}{"error": err.Error(), "client": c.uid, "user": c.UserID()}))
+		c.node.logger.log(newLogEntry(LogLevelInfo, "invalid subscription refresh token", map[string]interface{}{"error": errVerify.Error(), "client": c.uid, "user": c.UserID()}))
 		resp.Error = ErrorBadRequest
 		return resp, nil
 	}
 
-	if c.uid != tokenClient {
+	if c.uid != token.UserID {
```
FZambia: the same here: must be Client check

client.go
```diff
@@ -1835,94 +1743,58 @@ func (c *Client) subRefreshCmd(cmd *protocol.SubRefreshRequest) (*clientproto.Su
...
 
-	if len(tokenInfo) > 0 {
-		channelInfo = tokenInfo
-	}
-
-	if tokenB64info != "" {
-		byteInfo, err := base64.StdEncoding.DecodeString(tokenB64info)
-		if err != nil {
-			c.node.logger.log(newLogEntry(LogLevelInfo, "can not decode provided info from base64", map[string]interface{}{"user": c.UserID(), "client": c.uid, "error": err.Error()}))
-			return resp, DisconnectBadRequest
```
FZambia: previously we disconnected in case of info decode error

client_test.go
```diff
@@ -16,7 +16,7 @@ import (
 )
 
 func getConnToken(user string, exp int64) string {
-	claims := jwt.MapClaims{"user": user}
+	claims := jwt.MapClaims{"sub": user}
```
FZambia: good catch 👍

client.go
Nesty92: @FZambia UserID is Client string maybe we can rename that to avoid mistakes 
image

FZambia: Ah, yes - was not obvious, those are different things so `Client` is more correct anyway, thanks for fixing!

FZambia:
@Nesty92 rebased this into internal branch so I could fix some remaining things and simplify some behaviour regarding to error handling. So let's continue with #85 - feel free to review it and ask questions if sth is not obvious.

PR #498, opened 2025-07-25
Add history TTL configuration validation during message publishing 

yinheli:
## Summary

This PR implements history TTL configuration validation during message publishing to prevent runtime errors caused by incorrect TTL configurations.

## Background

This addresses #367, which was created in response to the Redis XADD errors reported in [centrifugal/centrifugo#768](https://github.com/centrifugal/centrifugo/issues/768)

Users encountered the following error when `history_ttl` was configured to be greater than `history_meta_ttl`:

### DIFF
--- a/errors.go
+++ b/errors.go
@@ -109,4 +109,10 @@ var (
 		Code:    112,
 		Message: "unrecoverable position",
 	}
+	// ErrorIncorrectHistoryTTLConfiguration means that history meta TTL is less than history TTL.
+	// History meta TTL must be greater than or equal to history TTL to avoid stream inconsistencies.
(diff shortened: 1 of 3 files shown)


FZambia:
Just realized I forgot to comment here - sorry @yinheli  for that.

I think the error here should not be part of client protocol (i.e. be an `*Error`) – it's more like an internal server error for clients. Some typed internal error is enough in the beginning I suppose.

It's not very handy for Centrifuge maybe to return such an error in runtime, but this is how library structured and the nature of ephemeral channels. In Centrifugo it's possible to validate configuration of retention policies on startup, so users won't hit such error in runtime.

PR #535, opened 2025-12-06
Further reduce allocs in write path

FZambia:
Given a bench:

```
go run main.go -s ws://localhost:8000/connection/websocket -n 10000 -ns 500 -np 100 -pl 10 -p channel
```

With changes introduced here we can see CPU reduction under batching scenario: 50% -> 36%. This further improves batching path after https://github.com/centrifugal/centrifuge/pull/530

In the path without active batching I did not notice significant changes in real CPU usage, but heap profile is all green, so the difference may be noticable under different tests:

Screenshot 2025-12-06 at 10 14 17 AM

UPD. This PR now includes https://github.com/centrifugal/centrifuge/pull/542 - see https://github.com/centrifugal/centrifuge/pull/535#issuecomment-3691684672

### DIFF
--- a/_examples/go.mod
+++ b/_examples/go.mod
@@ -33,7 +33,7 @@ require (
 	github.com/beorn7/perks v1.0.1 // indirect
 	github.com/bytedance/sonic v1.14.0 // indirect
 	github.com/bytedance/sonic/loader v0.3.0 // indirect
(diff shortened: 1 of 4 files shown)


FZambia:
This PR includes https://github.com/centrifugal/centrifuge/pull/542 now which comes with more optimizations of batching path. With `WriteWithTimer` option it's possible to eliminate writer goroutine and under some load profiles have reduced CPU utilization.

PR #541, opened 2025-12-18
Use hash tags for channels in cluster mode

FZambia:
Relates https://github.com/centrifugal/centrifugo/issues/1087 - and error `Script attempted to access keys that do not hash to the same slot script` while using Elasticache Serverless and publishing with history.

The root cause is described here: https://github.com/centrifugal/centrifugo/issues/1087#issuecomment-3667377731

Considering 2 approaches here:

1. Always use hash tags in cluster mode. This is simpler since server will work out of the box with every Redis setup. And it was actually a goal of Centrifuge/Centrifugo. But it will mean that during Centrifugo v6 release we will have incompatible changes in internal protocol - so nodes of different Centrifugo versions won't work together in one cluster. I guess it's not common at all and we never encouraged such setup, moreover combined with Redis Cluster (under 1% of setups) the chance of causing issues to someone is minimal. During rollout message loss can happen – but only on PUB/SUB layer, history streams and recovery logic will works as usual and will prevent client state corruptions.  
2. Adding a separate option to enable using hash tags for channels. While fully backward compatible will result into issues for every new user trying to setup Serverless Elasticache.

Now tend to go with option 1.

### DIFF
--- a/broker_redis.go
+++ b/broker_redis.go
@@ -1340,7 +1340,7 @@ func (b *RedisBroker) historyMetaKey(s *RedisShard, ch string) channelID {
 func (b *RedisBroker) extractChannel(isCluster bool, chID channelID) string {
 	ch := strings.TrimPrefix(string(chID), b.messagePrefix)
 
-	// Handle sharded pub/sub case: {idx}.channel
+	// Handle sharded PUB/SUB case: {idx}.channel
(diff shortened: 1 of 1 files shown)


FZambia:
Stick with option 1.

Issue #557, opened 2026-03-02
UpdatePresence using batch commit on addPresence


karasart:
**Is your feature request related to a problem? Please describe.**
No problem, just an optimize on efficiency.

**Describe the solution you'd like**
A new config field ClientPresenceUpdateBatchSize represents the batch size using, default 10s. Since lua ExecMulti will failed on redis-cluster, it will be check before use, and falls back to original if presence manager is cluster.

**Describe alternatives you've considered**
parallel on batch, but it seems checkSubscriptionExpiration function can't be that way.

**Additional context**
My collegues and i use centrifugal as our push service, and changes has been added to my company's code and it works great for 2 months, so i think if it's stable enough to submit a PR.

FZambia:
See https://github.com/centrifugal/centrifuge/pull/558#issuecomment-4414719014 :

> Let's close this one, but keep https://github.com/centrifugal/centrifuge/issues/557 open. I think that generally a scenario when a single client has a lot of channels with presence and Redis RTT starts affecting connection stability is worth investigating and addressing. But would be nice to do the work in parallel using a common pool of workers or configured per connection concurrency level and utilize pipelining.

FZambia:
One additional note here: we just introduced presence based on Map Subscriptions in https://github.com/centrifugal/centrifuge/pull/565 – so the work here should take it into account also (it's mostly the same as normal presence updates, just new periodic extra calls of `MapBroker` where presence is updated)

PR #558, opened 2026-03-04
Feat: UpdatePresence using batch commit on addPresence

karasart:
A new config field ClientPresenceUpdateBatchSize represents the batch size using, default by 1000. Since lua ExecMulti will failed on redis-cluster, it will be check before use, and falls back to original if presence manager is cluster. Should have some solid improvement in frequently sub/unsub situation.

# Benchmark：

**Feature:** Optimize Redis presence batch update with `AddPresenceBatch` + `ExecMulti`
**Environment:**

| Item | Value |
|------|-------|
| OS | macOS (Darwin 25.0.0, arm64) |
| CPU | Apple M5 (10 cores) |
| Go | 1.25.5 |
| Redis | 7-alpine (Docker, standalone, AOF/RDB off, maxmemory 512MB) |
| GOMAXPROCS | 10 |

---

## 1. Summary

Feature：Introduce `AddPresenceBatch`, which pipelines multiple Lua EVALSHA calls into a single `ExecMulti` round trip instead of issuing one network round trip per channel.

- The optimal batchSize lies in the **500 ~ 1000** range; larger values yield diminishing returns.
- At 10,000 channels: 
Serial:    **22.6x** faster.
Parallel: **5.7x** faster.

---

## 2. Large-Scale Parallel: 10,000 Channels

| Mode | batchSize | ms/op | vs Individual | B/op | allocs/op |
|------|----------:|------:|--------------:|---------:|----------:|
| Individual | - | 229.85 | baseline | 5,657,298 | 81,125 |
| Batch | 10 | 43.02 | **5.3x** | 9,345,332 | 100,471 |
| Batch | 50 | 40.77 | **5.6x** | 8,897,889 | 85,463 |
| Batch | 100 | 41.84 | **5.5x** | 8,825,383 | 83,494 |
| Batch | 500 | 42.53 | **5.4x** | 8,671,964 | 83,268 |
| Batch | 1000 | 40.49 | **5.7x** | 8,873,879 | 85,465 |
| Batch | 2500 | 41.59 | **5.5x** | 11,582,656 | 96,840 |
| Batch | 5000 | 46.16 | **5.0x** | 12,872,898 | 109,321 |
| Batch | 10000 | 44.48 | **5.2x** | 15,810,577 | 134,276 |

## 3. BatchSize Tuning: 10,000 Channels (Serial)

| batchSize | Batches | ms/op | vs Individual (1144ms) | B/op |
|----------:|--------:|------:|------------------------:|--------:|
| 100 | 100 | 80.10 | **14.3x** | 8,725,768 |
| 500 | 20 | 55.76 | **20.5x** | 8,386,328 |
| 1000 | 10 | 53.47 | **21.4x** | 8,315,624 |
| 2500 | 4 | 50.70 | **22.6x** | 9,016,848 |
| 5000 | 2 | 55.15 | **20.7x** | 9,049,708 |
| 10000 | 1 | 50.35 | **22.7x** | 9,486,396 |

> The optimal range is **batchSize=1000 ~ 2500**. At batchSize=5000 a slight regression appears (55ms vs 50ms), it came back at batchSize=10000 but 10000 is just too large for single time and it might cause redis's output buffer issue when processing other cmds.

## 4. Small-Scale Reference (Parallel)

| Channels | Individual (us/op) | Batch (us/op) |
|----------|-------------------:|--------------:|
| 1 | 5.4 | 13.0 |
| 5 | 49.0 | 46.6 |
| 10 | 47.7 | 52.1 |
| 50 | 262.9 | 198.0 |
| 100 | 444.9 | 397.2 |

## 5. Memory PresenceManager Baseline

In-memory presence manager, isolating pure data-structure cost from network overhead.

| Channels | Individual (ns/op) | Batch (ns/op) | Speedup |
|----------|-------------------:|--------------:|--------:|
| 1 | 23.35 | 23.19 | 1.0x |
| 10 | 318.4 | 238.3 | 1.3x |
| 50 | 1,624 | 1,188 | 1.4x |
| 100 | 3,532 | 2,311 | 1.5x |

> Batch is 25-35% faster even in-memory because it acquires the mutex once instead of N times.

---

## 6. Conclusions & Recommendations

1. **Batch could brings at least 5x faster**.

2. **Recommended batchSize: 500 above.**

### DIFF
--- a/client.go
+++ b/client.go
@@ -828,7 +828,7 @@ func (c *Client) updatePresence() {
 	// legacy per-channel path instead (handled by the caller via config check).
 	batchSize := config.ClientPresenceUpdateBatchSize
 	if batchSize <= 0 {
-		batchSize = 500 // Default to 500 channels per batch.
+		batchSize = 1000 // Default to 1000 channels per batch.
 	}
 
 	// Convert the channel map to a slice so it can be split into fixed-size batches.
--- a/client.go
+++ b/client.go
@@ -828,7 +828,7 @@ func (c *Client) updatePresence() {
 	// legacy per-channel path instead (handled by the caller via config check).
 	batchSize := config.ClientPresenceUpdateBatchSize
 	if batchSize <= 0 {
-		batchSize = 10 // Default to 10 channels per batch.
+		batchSize = 500 // Default to 500 channels per batch.
 	}
 
 	// Convert the channel map to a slice so it can be split into fixed-size batches.
--- a/client.go
+++ b/client.go
@@ -702,6 +702,50 @@ func (c *Client) updateChannelPresence(ch string, chCtx ChannelContext) error {
 	})
 }
 
+// updateChannelPresenceBatch updates client presence info for multiple channels in batch,
+// reducing Redis network round trips. This is more efficient than calling updateChannelPresence
+// multiple times.
+func (c *Client) updateChannelPresenceBatch(channels map[string]ChannelContext) error {
+	if len(channels) == 0 {
+		return nil
+	}
+
+	// Collect channels that need presence update
+	presenceBatchItems := make([]PresenceBatchItem, 0, len(channels))
+
+	c.mu.RLock()
+	for ch, chCtx := range channels {
+		// Skip channels without presence emission
+		if !channelHasFlag(chCtx.flags, flagEmitPresence) {
+			continue
+		}
+
+		// Verify channel still exists
+		if _, ok := c.channels[ch]; !ok {
+			continue
+		}
+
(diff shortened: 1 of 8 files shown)


FZambia:
Let's close this one, but keep https://github.com/centrifugal/centrifuge/issues/557 open. I think that generally a scenario when a single client has a lot of channels with presence and Redis RTT starts affecting connection stability is worth investigating and addressing. But would be nice to do the work in parallel using a common pool of workers or configured per connection concurrency level and utilize pipelining.

PR #591, opened 2026-07-24
Disable automatic read op retries in Redis

FZambia:
Centrifuge issues Redis commands without per-request context deadlines
and relies on IOTimeout to bound every call: when Redis is unreachable,
callers expect an error within roughly the IOTimeout envelope. Writes
and Lua scripts already behaved that way — they are not retried by
rueidis and fail on the first attempt.

Read-only commands did not. rueidis retries them on network errors
until the context is done, and with context.Background() that means
never: a presence-style read issued during a Redis outage kept
retrying (a burst of near-immediate attempts, then roughly once per
second) and simply did not return until Redis came back or the client
was closed.

Disable client-level retries entirely so reads fail the same way
writes do. Centrifuge surfaces errors to its callers, which own their
retry semantics; silently masking transient errors inside the client
is not a behavior anything relies on. Both the main and the replica
client inherit the setting since they share one ClientOption.

The test builds a shard through a local TCP proxy, stops the proxy to
produce the same errors a crashed Redis gives (broken connections,
refused dials), and requires a deadline-less GET to fail within
seconds. Without DisableRetry it retries against the dead address and
never returns.


FZambia:
Test race is fixed in #590

PR #601, opened 2026-07-31
Trim two per-publication costs on the hot path, unbreak the websocket benchmarks

FZambia:
Three independent changes found while profiling the fan-out and publish paths. The first two are behaviour-preserving refactors of hot-path plumbing; the third fixes benchmarks that could not complete.

## 1. Reuse the broadcast timestamp for per-subscriber position bookkeeping

`writePublicationUpdatePosition` stamped `channelContext.positionCheckTime` from its own `time.Now()` call — one clock read per subscriber per publication, for a value only ever stored and compared in whole seconds against a check delay measured in tens of them.

`subShard.broadcastPublication` already reads the clock once at the top, so that reading is threaded down to every subscriber. A fan-out completes in far less than the second of resolution the stamp carries, so the stored value is unchanged. Callers with no clock reading at hand pass `0` and the callee reads it itself.

Profiling a 1000-subscriber positioning fan-out put `time.Now` at **8.4% of total CPU**, all of it from this one function. Measured with base and patched binaries interleaved: **-16.4% on the broadcasting goroutine (p=0.000)** and **-3.9% process CPU per delivered message (p=0.035)**.

## 2. Skip the PublishOption closures on the client publish path

`Client.handlePublish` built a `[]PublishOption` holding `WithHistory`, `WithClientInfo` and sometimes `WithKey`, and handed them to `Node.Publish` — which immediately unwrapped them into a `PublishOptions` and discarded them. `PublishReply.Options` *is* a `PublishOptions`, so the closures existed only to be undone: four heap allocations per publication for a pure round trip.

The resolved path is split out as `Node.publishWithOptions` and the handler calls it directly. `Node.publish` still applies the options and delegates, so **`Node.Publish`, the `PublishOption` type and every `With*` constructor are untouched — no API change**.

The handler forwards the same explicit subset of fields it always has (history, client info, key). Passing `reply.Options` wholesale would compile and be faster still, but would silently start honouring `Tags`, `UseDelta`, `IdempotencyKey` and `Version` on a path that has never applied them.

Measured on the inbound publish command, interleaved: **14 -> 10 allocs/op, -19% B/op, -7.0% wall (p=0.002), -5.4% CPU (p=0.003)**.

## 3. Stop server pings from killing the websocket benchmarks mid-run

`BenchmarkWsPubSub`, the three `BenchmarkWsCommandReplyV2` variants and `BenchmarkWsBroadcastCompressionCache` each hold real websocket connections open across the measured loop, but their clients only read — they never answer an application-level ping. With the default 25s interval, any run long enough to reach it lost its connections and failed with `websocket: close 3012: no pong`.

At `-benchtime 1s`, four of eight sub-benchmarks failed and the parallel one failed every time; all pass now. The ping is pushed past any plausible run length in those five handlers. Tests in the file are short-lived and are left alone.

## Verification

- Full package test suite passes; `go build ./...` and `go vet ./...` clean.
- `gofmt -l` reports the same pre-existing files as `master` — none introduced here.
- Every A/B above compares two prebuilt test binaries run alternately. Running all of one side then all of the other gave the *opposite* sign on change 1 (a "significant" +6.6% regression) purely from ordering bias on a machine with ±15% run-to-run noise.

## Caveats

- All measurements are from a macOS laptop with both connection ends in one process. The directions are profile-attributed and solid; the magnitudes are not a substitute for a Linux run at scale.
- `time.Now` is a libc call on macOS and a vDSO call on Linux, so change 1's share will be smaller there.
- These are single-digit-percent wins. The large levers found in the same profiling session are configuration, not code: write batching (`WriteDelay`) is worth ~7-8x CPU per delivered message, and `UseWriteBufferPool` plus a smaller `ReadBufferSize` cut per-connection heap by ~34% and goroutines by half. Both default to off. Happy to open a separate discussion on defaults.
- The benchmarks used to produce these numbers are deliberately not included here.

### DIFF
--- a/client.go
+++ b/client.go
@@ -2563,17 +2563,16 @@ func (c *Client) handlePublish(req *protocol.PublishRequest, cmd *protocol.Comma
 		}
 
 		if reply.Result == nil {
-			publishOpts := []PublishOption{
-				WithHistory(reply.Options.HistorySize, reply.Options.HistoryTTL, reply.Options.HistoryMetaTTL),
-				WithClientInfo(reply.Options.ClientInfo),
-			}
-			if reply.Options.Key != "" {
-				publishOpts = append(publishOpts, WithKey(reply.Options.Key))
-			}
-			_, err := c.node.Publish(
-				event.Channel, event.Data,
-				publishOpts...,
-			)
+			// Forward exactly the options the closure-based form used to set -
+			// history, client info and key - and nothing else, so the rest of
+			// PublishReply.Options stays as unforwarded as it has always been.
+			_, err := c.node.publishWithOptions(event.Channel, event.Data, PublishOptions{
+				HistorySize:    reply.Options.HistorySize,
+				HistoryTTL:     reply.Options.HistoryTTL,
+				HistoryMetaTTL: reply.Options.HistoryMetaTTL,
+				ClientInfo:     reply.Options.ClientInfo,
+				Key:            reply.Options.Key,
+			})
 			if err != nil {
 				c.logWriteInternalErrorFlush(channel, protocol.FrameTypePublish, cmd, err, "error publish", started, rw)
 				return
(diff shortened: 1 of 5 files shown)


FZambia:
Superseded by #602, which carries only the benchmark fix. The two hot-path optimizations that were bundled in here are out of scope for a fix branch and can be proposed separately.

PR #628, opened 2026-09-13
Disconnect with DisconnectExpired when a connection refresh returns a past ExpireAt

FZambia:
## What
A client-side connection refresh handled an expired token in two different ways, depending on how the `OnRefresh` handler reported it:

| Refresh handler returns | Before | Client (centrifuge-js) |
|---|---|---|
| `RefreshReply{Expired: true}` | closes with `DisconnectExpired` (3005) | reconnects, gets a new token |
| `RefreshReply{ExpireAt: <past>}` | replies with `ErrorExpired` (110) | disconnects **without** reconnecting |

Both mean the connection's token has expired, but the second one left the client disconnected until the application called `connect()` again.

## Change
`handleRefresh` now closes the connection with `DisconnectExpired` when the new `ExpireAt` is not in the future, the same as for `Expired`. The client reconnects, and its connect with the old token gets `ErrorTokenExpired`, so it fetches a new token.

Handlers that set `Expired` are unaffected. Centrifugo sets `Expired` for an expired connection token, so this only changes behaviour for custom `OnRefresh` handlers that return a past `ExpireAt`.

## Tests
`TestClientSideRefreshExpired` covers two cases:
- **`Expired` flag:** connection closed with `DisconnectExpired`, no error reply
- **`ExpireAt` in the past:** same

Without the fix the `ExpireAt` in the past case fails: an `ErrorExpired` (110) reply is written. The `Expired` flag case pins behaviour that was already correct.

## Verification
- **Tests:** `go test -run Refresh .` passes.
- **Static checks:** `go vet .` and `gofmt` are clean.


FZambia:
Closing: an expired connection refresh is not an expected flow, keeping the current behaviour.

